PR #13, opened 2020-08-23
KTLN-133: Initializing Empty Mutable Collections in Kotlin

alimate:


kotlin-core/src/test/java/com/baeldung/mutablecollection/MutableCollectionUnitTest.kt
```diff
@@ -0,0 +1,26 @@
+package com.baeldung.mutablecollection
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class MutableCollectionUnitTest {
```
KevinGilmore, line 6: Shouldn't Kotlin test files go in src/test/kotlin?

PR #14, opened 2020-08-23
KTLN-163: Static initialisation block in Kotlin

alimate:


kotlin-core/src/main/kotlin/com/baeldung/staticinit/JavaStatic.java
```diff
@@ -0,0 +1,8 @@
+package com.baeldung.staticinit;
+
+public class JavaStatic {
+
```
KevinGilmore, line 4: Let's omit the Java class from the PR.

PR #51, opened 2021-02-26
Add securityMatcher to AdminSecurityConfiguration

eleftherias:
Without a custom `securityMatcher`, the `AdminSecurityConfiguration` will match all requests, meaning no request will reach `BasicSecurityConfiguration`.

Setting the `securityMatcher` to `"/greetings/**"` means requests that match `"/greetings/**"` will be processed by the `AdminSecurityConfiguration`, while the rest of the requests will be processed by the `BasicSecurityConfiguration`.
This is the behaviour that the tutorial describes.

The code should also be updated in the tutorial https://www.baeldung.com/kotlin/spring-security-dsl


lor6:
Thanks @eleftherias I've updated the tutorial.

PR #62, opened 2021-05-08
KTLN-286: Filtering a list in Kotlin

anirban99:


glmartin (changes requested):
No need to include changes to the README. We handle that when the article is published.

anirban99:
> No need to include changes to the README. We handle that when the article is published.

Removed ReadMe

glmartin:
The author is on hold for a few months.

PR #91, opened 2021-07-14
KTLN-306: Creating a new instance of a KClass

alimate:


core-kotlin-modules/core-kotlin-advanced/src/test/kotlin/com/baeldung/inatantiation/ReflectiveInstantiationUnitTest.kt
```diff
@@ -0,0 +1,59 @@
+package com.baeldung.inatantiation
```
eric-martin, line 1: Package name typo (inatantiation)

PR #114, opened 2021-08-05
KTLN-231: Returning multiple values from a function in Kotlin

alimate:


core-kotlin-modules/core-kotlin-advanced/src/main/kotlin/com/baeldung/returnmultiple/main.kt
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.returnmultiple
+
+import java.net.InetAddress
+import java.security.KeyPair
+import java.security.KeyPairGenerator
+
+fun twoPair(): Pair<String, Int> = "Ali" to 33
+fun threeValues(): Triple<String, Int, String> = Triple("Ali", 33, "Neka")
+fun fourVales() = arrayOf("Berlin", "Munich", "Amsterdam", "Madrid", "Vienna")
```
KevinGilmore, line 9: looks like five values

KevinGilmore (commented):
The example "fourValues" actually uses five values.

PR #151, opened 2021-11-02
KTLN-134: Abstract Factory Pattern in Kotlin

diegotorrespy:


core-kotlin-modules/core-kotlin-design-patterns/src/test/kotlin/com/baeldung/abstractfactory/AbstractFactoryPatternUnitTest.kt
```diff
@@ -0,0 +1,30 @@
+package com.baeldung.abstractfactory
+
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+internal class AbstractFactoryPatternUnitTest {
+
+    @Test
+    fun crossbowWeaponFactoryTest() {
```
glmartin, line 10: Let's change the test method names to be in the format of given_When_Then(). For example `givenCrossbowFactory_whenBuildWeapon_thenCrossbowIsReturned()`.

PR #234, opened 2022-05-03
Updated README.md

Asjad-J:
added link back to the article: https://www.baeldung.com/kotlin/call-function-after-delay


Asjad-J:
test this please

PR #237, opened 2022-05-05
[KTLN-148] Add examples for Kotlin Flow

Sulion:


Sulion:
Fails due to some "jee-kotlin". I have no idea what it is.

PR #248, opened 2022-05-29
Use companion object

behnamkhani-git:


core-kotlin-modules/core-kotlin-lang-oop/src/main/kotlin/com/baeldung/companionobject/CompanionObject.kt
```diff
@@ -0,0 +1,62 @@
+package com.baeldung.companionobject
+
+class CompanionObject {
+    class ClassName {
+        fun anotherFun() {
+            println(propertyName)
+        }
+
+        companion object {
+            const val propertyName: String = "Something..."
```
Bwvolleyball, line 10: Let's annotate this property with `@JvmStatic` like the document:

```suggestion
        companion object {
            // companion object members can be marked @JvmStatic for better java inter-op
            @JvmStatic
            const val propertyName: String = "Something..."
```

core-kotlin-modules/core-kotlin-lang-oop/src/main/kotlin/com/baeldung/companionobject/CompanionObject.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    class MyClass {
+        companion object Factory {
+            fun createInstance(): MyClass = MyClass()
+        }
+    }
+
+
```
Bwvolleyball, line 23: Let's add the interface with a companion object here too:

```suggestion
    class MyClass {
        companion object Factory {
            fun createInstance(): MyClass = MyClass()
        }
    }

    interface MyInterface { 
        companion object { 
            const val PROPERTY = "value"
        }
    }

```

core-kotlin-modules/core-kotlin-lang-oop/src/main/kotlin/com/baeldung/companionobject/CompanionObject.kt
behnamkhani-git: ok, now propertyName is equipped with @JvmStatic.

Bwvolleyball (approved):

PR #266, opened 2022-06-18
KTLN-13 sort with multiple fields

ryan728:


Bwvolleyball (commented):
PR looks great, but lets move this over to [core-kotlin-collections-3](https://github.com/Baeldung/kotlin-tutorials/tree/master/core-kotlin-modules/core-kotlin-collections-3) since I think this module is a bit full

PR #279, opened 2022-06-28
KTLN-223 Constants best practices

polomos:


core-kotlin-modules/core-kotlin-lang-3/src/main/kotlin/com/baeldung/constants/ConstantsBestPractices.kt
```diff
@@ -0,0 +1,10 @@
+package com.baeldung.constants
+
+//const val constantAtTopLevel = SimpleClass()
```
Bwvolleyball, line 3: Let's add a comment here that this won't compile since it's not a primitive or string.

```suggestion

// This won't compile since SimpleClass is not a primitive or String.
// const val constantAtTopLevel = SimpleClass()
```

PR #295, opened 2022-07-31
[KTLN-259] default non optional param

LazyBun:


Bwvolleyball:
I'm wondering if this tutorial would be better placed in this [core-modules-5](https://github.com/Baeldung/kotlin-tutorials/tree/master/core-kotlin-modules/core-kotlin-5) folder?

LazyBun:
I'm fine with moving this around - I did not have a great idea on where to place it. Will move.

PR #330, opened 2022-09-30
KTLN-382: Unit tests for Removing Whitespaces

diegotorrespy:
KTLN-382: Unit tests for Removing Whitespaces


core-kotlin-modules/core-kotlin-strings-2/src/test/kotlin/com/baeldung/whitespaces/RemoveWhitespaceTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        val withOutSpaces = example.replace("\\s".toRegex(), "")
+
+        assertThat(withOutSpaces).hasToString("Kotlinisaprogramminglanguage")
+    }
+
+    @Test
+    fun givenString_whenFilterByNonWhitespace_WhitespacesShouldBeRemoved() {
+
+        val example = "House Of The Dragon"
+        val withOutSpaces = example.filter { !it.isWhitespace() }
```
Bwvolleyball, line 30: Mentioned this in the feedback as well, but maybe

```suggestion
        val withOutSpaces = example.filterNot { it.isWhitespace() }
```

would be clearer?

core-kotlin-modules/core-kotlin-strings-3/README.md
```diff
@@ -0,0 +1,7 @@
+## Core Kotlin Strings
+
+This module contains articles about core Kotlin strings.
+
+### Relevant articles:
+
+- [Removing Whitespace from Strings](https://www.baeldung.com)
```
Bwvolleyball, line 7: I think we can leave this off for now and it'll be added back later when the article is processed.

```suggestion
```

PR #365, opened 2022-11-29
KTLN-391: Difference between two dates in Kotlin

nrsuresh:
https://drafts.baeldung.com/kotlin/wp-admin/post.php?post=53493&action=edit


core-kotlin-modules/core-kotlin-date-time/src/main/kotlin/com/baeldung/dates/datetime/DateDifference.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        val years = period.years
+        val months = period.months
+        val days = period.days
+
+        return mapOf(
+            Calendar.YEAR to years,
+            Calendar.MONTH to months,
+            Calendar.DATE to days
+        )
```
JavierPeris, line 27: As a consumer of a diff between two dates I would expect a Period object instead of a map.

JavierPeris (approved):

nrsuresh:
Updated the code per article review comments

PR #379, opened 2022-12-18
Jira Issue: KTLN-412, ewaneenoombe@gmail.com

nappy29:
Changes are made to the core-kotlin-strings-3 module.
It contains test cases as applied to various extension functions provided in Kotlin library for the String class.


core-kotlin-modules/core-kotlin-io/src/main/kotlin/com/baeldung/files/creation/FileCreator.kt
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.files.creation
+
+import java.io.File
+
+class FileCreator {
```
Bwvolleyball, line 5: What's the goal of this? It feels extra to what the article is about?

core-kotlin-modules/core-kotlin-strings-3/src/main/kotlin/com/baeldung/string/extension_functions/StringExtensionFunctionClass.kt
```diff
@@ -0,0 +1,37 @@
+package com.baeldung.string.extension_functions
+
+import java.nio.charset.Charset
+
+class StringExtensionFunctionClass {
```
Bwvolleyball, line 5: Hmm, it feels weird to wrap the examples in this 'helper' class - why not just use the extension functions directly in the test?

core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/string_extension_functions/StringExtensionFunctionClassUnitTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+
+    @Test
+    fun `using replace() method`(){
+        val inputString0 = "Jelly"
+        val inputString1 = "Kotlin Replace Program"
+        assertEquals(inputString0.replace( 'l', 'z'), "Jezzy")
+        assertEquals(inputString0.replace('j', 'P', true), "Pelly")
+        assertEquals(inputString0.replace('j', 'P'), "Jelly")
+        assertEquals(inputString1.replace("PROGRAM", "Examples", true), "Kotlin Replace Examples")
```
KevinGilmore, line 17: One these and all other assertEquals calls in this file, the order of arguments need to be swapped, so that you end up with assertEquals(expected, actual).

core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/string_extension_functions/StringExtensionFunctionClassUnitTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        var str = "Extension Functions"
+        var result = str.uppercase()
+        assertEquals("EXTENSION FUNCTIONS",result)
+    }
+
+    @Test
+    fun `using lowercase() method`(){
+        var str = "Extension Functions"
+        var result = str.lowercase()
+        assertEquals(result, "extension functions")
```
KevinGilmore, line 31: need to swap the args

nappy29: That has been fixed now.
Thank you for pointing it out.

PR #405, opened 2023-02-10
[KTLN-235] Add InappropriateBlocking

LeoColman:


core-kotlin-modules/core-kotlin-6/README.md
```diff
@@ -7,3 +7,4 @@
 - [How to Invoke External Command From Within Kotlin](https://www.baeldung.com/kotlin/invoke-external-command)
 - [Kotlin Equivalent to Java String Arrays](https://www.baeldung.com/kotlin/java-string-arrays-equivalent)
 - [Kotlin Bit Operations](https://www.baeldung.com/kotlin/bit-operations)
+- [Kotlin Error "Accidental override: The following declarations have the same JVM signature"]()
```
Bwvolleyball, line 10: Let's delete this line - it'll be added back in after this article is published, and will fix the conflicts.

```suggestion
```

core-kotlin-modules/core-kotlin-6/src/main/kotlin/com/baeldung/accidental_override/Animals.kt
```diff
@@ -0,0 +1,24 @@
+package com.baeldung.accidental_override
+
+//class Cat : Animal {
+//    val sound: String = "Meow"
+//}
```
Bwvolleyball, line 5: Should we add a comment that explains the error this would create?

```suggestion
// This example will create an accidental override error, because of the conflicting `getSound()` function.
//class Cat : Animal {
//    val sound: String = "Meow"
//}
```

PR #407, opened 2023-02-13
KTLN-338 Singleton class in Kotlin

AndiCover:
https://jira.baeldung.com/browse/KTLN-338


JavierPeris (approved):
The core-kotlin-3 module is full (it already has 10 articles). So you should use kotlin-6-module instead.

JavierPeris (approved):

PR #415, opened 2023-02-19
Code for KTLN-398

vinaywadhwa:


core-kotlin-modules/core-kotlin-concurrency-2/src/main/kotlin/com/baeldung/coroutine/dispatchers/IoAndDefaultDemo.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+import org.json.JSONObject
+
+class IoAndDefaultDemo {
+    private val ioDispatcher = Dispatchers.IO
+    private val defaultDispatcher = Dispatchers.Default
```
Bwvolleyball, line 11: Do these already reference objects, or constants? If so, would it be more powerful to just use `Dispatchers.IO` and `Dispatchers.Default` in-line respectively, instead of storing these as extra variables?

PR #424, opened 2023-03-01
[KTLN-422] Add code for RunCatching and Try Finally

LeoColman:


core-kotlin-modules/core-kotlin-6/src/main/kotlin/com/baeldung/runcatching_tryfinally/RunCatchingTryFinally.kt
```diff
@@ -0,0 +1,28 @@
+package com.baeldung.runcatching_tryfinally
```
Bwvolleyball, line 1: This file is a duplicate, and left-over from moving out of core-kotlin-6, let's delete it.

PR #434, opened 2023-03-10
Issue No: KTLN-228; email:ewaneenombe@gmail.com

nappy29:
Included test cases for usage of Lombok in Kotlin


core-kotlin-modules/core-kotlin-annotations/src/test/kotlin/com/baeldung/lombok/usage/KotlinLombokConfiguration.kt
```diff
@@ -0,0 +1,60 @@
...
+        p1.age = 20
+        p1.gender = "Male"
+
+
+        assertEquals("Nappy", p.name)
+        assertEquals(20, p.age)
+        assertEquals("Male", p.gender)
+
+        assertTrue(p == p1)
+        assertTrue(p.hashCode() == p1.hashCode())
```
ashleyfrieze, line 57: Ideally we'd use `assertEquals` here, rather than `assertTrue`. If we're using `==` in an assertion than all assertions could be replaced by `assertTrue(<expr>)` which isn't how we use assertions

nappy29: Okay got it. 
Made the fixes as asked

PR #450, opened 2023-04-05
[KTLN-397] Kotlin Coroutines with timeout

andrsim66:


core-kotlin-modules/core-kotlin-concurrency-2/src/test/kotlin/com/baeldung/coroutine/timeouts/TimeoutAndResourceDemoTest.kt
```diff
@@ -0,0 +1,20 @@
...
+            acquireAndReleaseWithLeak()
+        }
+    }
+
+    @Test
+    fun when_resource_accessed_by_reference_inside_timeout_block_then_acquired_value_is_always_0() {
+        repeat(100) {
+            acquireAndReleaseWithoutLeak()
+        }
+    }
```
Bwvolleyball, line 19: A few questions here:

1. can we make these functions return something so that there's something to test on?
2. Can we re-work these to be more deterministic examples, and therefore easier to test?

For both tests, we could flip the timeout to be 50, and the delay to be longer (60), so that we always hit the timeout. In that way, we can prove that the first example never releases resources, and the second one always does, without needing to repeat them 100 times, in that example, once would be enough.

core-kotlin-modules/core-kotlin-concurrency-2/src/test/kotlin/com/baeldung/coroutine/timeouts/TimeoutsDemoTest.kt
Bwvolleyball, line 1: It feels like these tests should be asserting on something being returned as well.

PR #462, opened 2023-04-22
[KTLN-405] Add code samples to Generic Inline Tests

LeoColman:


core-kotlin-modules/core-kotlin-advanced-2/README.md
```diff
@@ -8,3 +8,4 @@ This module contains articles about advanced topics in Kotlin.
 - [Get the Type of a Variable in Kotlin](https://www.baeldung.com/kotlin/find-type-variable)
 - [Serialize Kotlin Data Class With Default Values Into JSON](https://www.baeldung.com/kotlin/data-class-json-serialize-default-values)
 - [How to Serialize Arrays in Kotlin With Kotlinx.serialization](https://www.baeldung.com/kotlin/kotlinx-serialization)
+- [Kotlin Generic Inline Classes]()
```
Bwvolleyball, line 11: It's safe to leave this off, as the core team will add this when they process the article after publication.

core-kotlin-modules/core-kotlin-advanced-2/src/main/kotlin/com/baeldung/generic/inline/Samples.kt
```diff
@@ -0,0 +1,46 @@
...
+@JvmInline
+value class Box<T>(val value: T)
+
+val intBox: Box<Int> = Box(42)
+val stringBox: Box<String> = Box("Hello, Baeldung!")
+
+
+@JvmInline
+value class Name<T : CharSequence>(val value: T) {
+    operator fun plus(other: Name<*>): Name<CharSequence> = Name(value.toString() + " " + other.value)
```
Bwvolleyball, line 45:
```suggestion
    operator fun plus(other: Name<T>): Name<T> = Name("$value ${other.value}")
```

core-kotlin-modules/core-kotlin-advanced-2/src/main/kotlin/com/baeldung/generic/inline/Samples.kt
```diff
@@ -0,0 +1,46 @@
...
+    val id: Id<T>
+}
+
+data class User(override val id: Id<String>, val name: String) : Identifiable<String>
+
+data class Wallet(override val id: Id<Int>, val currency: String) : Identifiable<Int>
+
+fun <T : Identifiable<*>> findById(entities: List<T>, id: Id<*>): T? {
+    return entities.firstOrNull { it.id == id }
+}
```
Bwvolleyball, line 26: Thinking out loud here, would this read better?

```suggestion
fun <T : Identifiable<ID: Any>> findById(entities: List<T>, id: Id<ID>): T? {
    return entities.firstOrNull { it.id == id }
}
```

LeoColman: I think it is clearer, yes!
In production I would probably leave the wildcard to mean "I'm leaving this as I intend to not care about the type". For tutorial reasons (and for this, specially) we totally care about the type xD

PR #475, opened 2023-05-06
[KTLN-506] Add samples

LeoColman:


core-kotlin-modules/core-kotlin-7/src/main/kotlin/com/baeldung/mapfunction/Samples.kt
```diff
@@ -0,0 +1,27 @@
...
+    this[i] = transform(this[i])
+  }
+  return this
+}
+
+fun useMapInPlace() {
+  val numbers = arrayOf(1, 2, 3, 4, 5)
+  numbers.mapInPlace { it * 2 }
+  println(numbers) // [2, 4, 6, 8, 10]
+}
```
Bwvolleyball, line 27: I'm wondering if these samples would be better off as some unit tests, and instead of println's, we just assert equality to the commented list / boolean

core-kotlin-modules/core-kotlin-7/src/main/kotlin/com/baeldung/mapfunction/mapInPlace.kt
```diff
@@ -0,0 +1,8 @@
+package com.baeldung.mapfunction
+
+fun <T> Array<T>.mapInPlace(transform: (T) -> T): Array<T> {
```
KevinGilmore, line 3: Pretty sure Kotlin's standard indent is 4 spaces per level. Let's fix that.

PR #489, opened 2023-05-18
[KTLN-515] Add samples

LeoColman:


core-kotlin-modules/core-kotlin-7/src/main/kotlin/com/baeldung/optional/Optional.kt
```diff
@@ -0,0 +1,33 @@
...
+data class User(val name: String)
+
+fun getUsername(user: User?): String {
+    val optionalUser = Optional.ofNullable<User>(user)
+    return optionalUser.map(User::name).orElse("Unknown")
+}
+
+fun getUsernameKotlin(user: User?): String {
+    return user?.name ?: "Unknown"
+}
```
Bwvolleyball, line 33: Let's actually turn this example code into a unit test.  The code in the article is fine though. For this, since it's optional unwrapping and nullability, I think just asserting null vs non null is enough.

core-kotlin-modules/core-kotlin-7/src/main/kotlin/com/baeldung/optional/Optional.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlin.jvm.optionals.getOrNull
+
+fun getOrNull() {
+    val optionalValue: Optional<String> = Optional.of("Hello, Baeldung!")
+    val value: String? = optionalValue.getOrNull()
+}
+
+fun getOrElse() {
+    val optionalValue: Optional<String> = Optional.empty()
+    val value: String = optionalValue.getOrElse { throw IllegalStateException("Value is missing") }
```
Bwvolleyball, line 15: This one we should wrap with the `assertThatThrows()` or equivalent when we convert to a unit test.

core-kotlin-modules/core-kotlin-7/src/main/kotlin/com/baeldung/optional/Optional.kt
```diff
@@ -0,0 +1,33 @@
...
+    val value2: String = optionalValue.getOrElse { "Default Value" }
+}
+
+fun getOrDefault() {
+    val optionalValue: Optional<String> = Optional.empty()
+    val value1: String = optionalValue.getOrDefault("Default Value")
+}
```
Bwvolleyball, line 22: Let's also actually delete these 3 functions, since the test is a better demonstration of usage and so that we don't duplicate the code

Bwvolleyball:
```suggestion
```

core-kotlin-modules/core-kotlin-7/src/test/kotlin/com/baeldung/optional/OptionalTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.api.assertThrows
+import java.util.Optional
+import kotlin.jvm.optionals.getOrDefault
+import kotlin.jvm.optionals.getOrElse
+import kotlin.jvm.optionals.getOrNull
+import kotlin.test.assertEquals
+
+class OptionalTest {
+
+  @Test
```
KevinGilmore, line 13: Standard indent for Kotlin is 4 spaces per level

PR #542, opened 2023-07-09
[KTLN-559] Add samples

LeoColman:


kotest/src/main/kotlin/com/baeldung/kotest/beforetest/Database.kt
```diff
@@ -0,0 +1,14 @@
+package com.baeldung.kotest.beforetest
+
+
+data class User(val id: Long, val username: String)
+
+class Database {
```
Bwvolleyball, line 6: Stylelistic, but would it make more sense to call this the `UserService` or `UserRepository` class?

kotest/src/test/kotlin/com/baeldung/kotest/beforetest/BeforeTestSamples.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class BeforeTestSamples : FunSpec({
+
+    val userRepository = UserRepository()
+
+    beforeEach {
+        userRepository.addUser(User(1, "Admin"))
+    }
+
+    isolationMode = IsolationMode.InstancePerTest
```
Bwvolleyball, line 17: Let's flip the order of these.

```suggestion
    isolationMode = IsolationMode.InstancePerTest

    beforeEach {
        userRepository.addUser(User(1, "Admin"))
    }
```

PR #572, opened 2023-08-08
KTLN-632: Check if a String Is a Palindrome in Kotlin

nappy29:
Added Unit test

### DIFF
--- a/core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/palindromicstring/PalindromicStringTest.kt
+++ b/core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/palindromicstring/PalindromicStringTest.kt
@@ -1,0 +1,61 @@
+package com.baeldung.palindromicstring
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+class PalindromicStringTest {
+
+    @Test fun `palindrome check using loop function`(){
+        assert(isPalindrome("racecar") == true)
+        assert(isPalindrome("hello") == false)
+        assert(isPalindrome("level") == true)
+        assert(isPalindrome("madam") == true)
+        assert(isPalindrome("3553") == true)
+        assert(isPalindrome("123213") == false)
+    }
+
+    @Test fun `palindrome check using built in function`(){
+        assert(isPalindromeWithBuiltInFunction("racecar") == true)
+        assert(isPalindromeWithBuiltInFunction("dadar") == false)
+        assert(isPalindromeWithBuiltInFunction("edrarde") == true)
+        assert(isPalindromeWithBuiltInFunction("radar") == true)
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/palindromicstring/PalindromicStringTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    fun isPalindromeUsingRecursion(str: String): Boolean {
+        if (str.length <= 1) {
+            return true
+        }
+        if (str.first() != str.last()) {
+            return false
+        }
+        return isPalindrome(str.substring(1, str.length - 1))
```
Bwvolleyball, line 59: Looks like this should be `isPalindromeUsingRecursion` instead.

```suggestion
        return isPalindromeUsingRecursion(str.substring(1, str.length - 1))
```

core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/palindromicstring/PalindromicStringTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        var j = str.length - 1
+        while (i < j) {
+            if (str[i] != str[j]) {
+                return false
+            }
+            i++
+            j--
+        }
+        return true
+    }
```
Bwvolleyball, line 45:
```suggestion
    fun isPalindrome(str: String): Boolean {
        var start = 0
        var end = str.length - 1
        while (start < end) {
            if (str[start] != str[end]) {
                return false
            }
            start++
            end--
        }
        return true
    }
```

nappy29: Got It Sir.

nappy29: Got It Sir.

core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/palindromicstring/PalindromicStringTest.kt
```diff
@@ -0,0 +1,60 @@
+package com.baeldung.palindromicstring
+import org.junit.jupiter.api.Test
+
+class PalindromicStringTest {
+
+    @Test fun `palindrome check using loop function`(){
+        assert(isPalindromeUsingLoop("racecar") == true)
```
KevinGilmore, line 7: For all trivial assertions, prefer either the JUnit5's assertTrue, assertFalse, etc, or use the AssertJ library's fluent assertions.

PR #586, opened 2023-08-20
[KTLN-627] Add samples

LeoColman:


kotest/src/main/kotlin/com/baeldung/kotest/datadriven/isPythagTriple.kt
Bwvolleyball, line 1: Let's make this file name start with an upper case letter at a minimum.

Should it be called `PythagTriple`, `IsPythagTriple`, `PythagTripleUtil` etc?

PR #588, opened 2023-08-21
KTLN-646: update khttp dependency location

dariusandz:


kotlin-libraries-http/pom.xml
```diff
@@ -14,9 +14,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>khttp</groupId>
+            <groupId>org.danilopianini</groupId>
             <artifactId>khttp</artifactId>
-            <version>${khttp.version}</version>
+            <version>1.3.1</version>
```
Bwvolleyball, line 19: Let's actually keep this version as a property in the pom.xml

dariusandz: Done

PR #589, opened 2023-08-22
KTLN-631: Check Whether a Character is a Vowel or Consonant in Kotlin

nappy29:
Added unit test

### DIFF
--- a/core-kotlin-modules/core-kotlin-8/src/test/kotlin/com/baeldung/vowelOrConsonant/VowelOrConsonantUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-8/src/test/kotlin/com/baeldung/vowelOrConsonant/VowelOrConsonantUnitTest.kt
@@ -82,8 +82,8 @@ class VowelOrConsonantUnitTest {
     }
 
     fun isConsonantUsingRegex(c: Char): Boolean {
-        val consonants = "[bcdfghjklmnpqrstvwxyz]"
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-8/src/test/kotlin/com/baeldung/vowelOrConsonant/VowelOrConsonantUnitTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    fun isVowelUsingAsciiValues(c: Char): Boolean {
+        val ascii = c.code
+        return ascii == 65 || ascii == 69 || ascii == 73 || ascii == 79 || ascii == 85 || ascii == 97 || ascii == 101 || ascii == 105 || ascii == 111 || ascii == 117
+    }
+
+    fun isConsonantUsingAscii(c: Char): Boolean {
+        val ascii = c.lowercaseChar().code
+        return ascii in 97..122 && !isVowelUsingAsciiValues(c)
```
Bwvolleyball, line 96: Let's actually just show this handling both upper case and lowercase directly, since it's valuable to see the split range of ascii uppercase and lowercase char ranges.

```suggestion
        val ascii = c.code
        return (ascii in 65..90 || ascii in 97..122) && !isVowelUsingAsciiValues(c)
```

PR #596, opened 2023-08-30
code for - KTLN-616: Comparator in Kotlin

vinaywadhwa:


core-kotlin-modules/core-kotlin-8/src/test/kotlin/com/baeldung/comparator/ComparatorExample.kt
```diff
@@ -0,0 +1,68 @@
...
+    fun `when custom comparator used to sort list then list sorted correctly`() {
+        val words = listOf("apple", "blueberry", "cherry", "date")
+        val lengthComparator = object : Comparator<String> {
+            override fun compare(string1: String, string2: String): Int {
+                // return 0 if strings are of equal length, a negative number if the first string is shorter than the second one, or a positive number otherwise
+                return string1.length - string2.length
+            }
+        }
+        assertIterableEquals(words.sortedWith(lengthComparator), listOf("date", "apple", "cherry", "blueberry"))
+        assertIterableEquals(words.sortedWith(compareBy(String::length)), listOf("date", "apple", "cherry", "blueberry"))
```
KevinGilmore, line 18: In all the assertIterableEquals calls, the expected value should be the first argument, and the actual value should be the second argument.

PR #606, opened 2023-09-07
KTLN-655: Kotlinx Serialization Inheritance

dariusandz:


kotlin-libraries-data/src/test/kotlin/com/baeldung/serialization/kotlinx/polymorphism/ClosedPolymorphismSerializationUnitTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+@Serializable
+sealed class Shape
```
Bwvolleyball, line 11: Don't forget to add the extra base property to this object here as well.

dariusandz: Done

PR #609, opened 2023-09-10
Insertion sort

kanake10:


core-kotlin-modules/core-kotlin-datastructures/src/main/kotlin/com/baeldung/insertionsort/SortArray.kt
```diff
@@ -0,0 +1,27 @@
...
+    val people = arrayOf(
+        Person("Alice", 30),
+        Person("Bob", 25),
+        Person("Charlie", 35),
+        Person("David", 28)
+    )
+    insertionSortByAge(people)
+    for (person in people) {
+        println("${person.name}, ${person.age}")
+    }
```
Bwvolleyball, line 26: The snippets in the draft make this look like test assertions, should we actually call this sort in a test instead and verify which elements end up in which position?

core-kotlin-modules/core-kotlin-datastructures/src/main/kotlin/com/baeldung/insertionsort/SortIntArray.kt
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
+
+fun insertionSortWithIntArray() {
+    val myArray = intArrayOf(5, 2, 9, 3, 6)
+    insertionSort(myArray)
+    println("Sorted Array:")
+    for (element in myArray) {
+        println(element)
+    }
```
Bwvolleyball, line 20: Same with this part, it feels like we should have a test that is asserting the final order of the now sorted array.

PR #642, opened 2023-10-04
Update insertion

kanake10:


core-kotlin-modules/core-kotlin-datastructures/src/test/kotlin/com/baeldung/insertionsort/SortArrayTest.kt
```diff
@@ -14,7 +14,7 @@ fun insertionSortByAge(people: Array<Person>) {
         val key = people[i]
         var j = i - 1
 
-        while (j >= 0 && people[j].age > key.age) {
+        while (j >= 0 && people[j].compareTo(key) > 0) {
```
Bwvolleyball, line 17: Let's update this PR as well to be:
```suggestion
        while (j >= 0 && people[j] > people[i]) {
```

PR #644, opened 2023-10-07
KTLN-706: Create a Frequency Map in Kotlin

nappy29:
added unit tests

### DIFF
--- a/core-kotlin-modules/core-kotlin-datastructures/src/test/kotlin/com/baeldung/frequencymap/FrequencyMapUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-datastructures/src/test/kotlin/com/baeldung/frequencymap/FrequencyMapUnitTest.kt
@@ -14,7 +14,7 @@ class FrequencyMapUnitTest {
     }
 
     @Test
-    fun `test frequency map using groupBy() method`() {
+    fun `test frequency map using groupingBy() method`() {
         val list = listOf(1, 2, 1, 3, 2, 4, 4, 7,9,7,3,2,1)
         val expectedMap = mapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
         val actualMap = list.groupingBy { it }.eachCount()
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-datastructures/src/test/kotlin/com/baeldung/frequencymap/FrequencyMapUnitTest.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    return map
+}
+
+fun frequencyMapUsingBinarySearchMethod(list: List<Int>): MutableMap<Int, Int>{
+    val sortedList = list.sorted()
+
+    val map = mutableMapOf<Int, Int>()
+
+    sortedList.forEach { element ->
```
Bwvolleyball, line 97: One small optimization here:

```suggestion
    sortedList.distinct().forEach { element ->
```

nappy29: Thanks for pointing that out.
Fixed!

core-kotlin-modules/core-kotlin-datastructures/src/test/kotlin/com/baeldung/frequencymap/FrequencyMapUnitTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Test
+import java.util.*
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+
+class FrequencyMapUnitTest {
+
+    @Test
+    fun `test frequency map using mutable map method`() {
+        val list = listOf(1, 2, 1, 3, 2, 4, 4, 7,9,7,3,2,1)
```
KevinGilmore, line 12:
```suggestion
        val list = listOf(1, 2, 1, 3, 2, 4, 4, 7, 9, 7, 3, 2, 1)
```

KevinGilmore: We need to add spaces here for readability. Let's apply this change in the other tests, too.

PR #649, opened 2023-10-11
string_ iteration

kanake10:


core-kotlin-modules/core-kotlin-strings-3/src/test/kotlin/com/baeldung/backwardIteration/IterationTest.kt
```diff
@@ -0,0 +1,30 @@
...
+        }
+        return reversedText.toString()
+    }
+
+    @Test
+    fun testIterateStringWithForLoop() {
+        val mock = mockk<IterationTest>()
+        every { mock.iterateStringWithForLoop() } returns "nittoK ,olleH"
+        val result = mock.iterateStringWithForLoop()
+        verify { mock.iterateStringWithForLoop() }
```
Bwvolleyball, line 25: Why mock this? Why not just call the live function?

PR #659, opened 2023-10-19
KTLN-585 Expedia Group GraphQL Kotlin

VladGaluska:


kotlin-ktor/gradlew
theangrydev, line 1: I had issues running this and had to:
```
gradle wrapper
chmod +x gradlew
```

To be able to run it without error. Can you please check and recommit the corrected wrapper files?

kotlin-ktor/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
+import com.expediagroup.graphql.plugin.gradle.config.GraphQLSerializer
+
+description = "Example usage of Gradle plugin to generate GraphQL Kotlin Client"
+
+val expediaGroupVersion = "7.0.1"
```
theangrydev, line 5: Rename to match the actual library name GraphQL Kotlin, something like:
```suggestion
val graphQLKotlinVersion = "7.0.1"
```

kotlin-ktor/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+    id("io.ktor.plugin") version "2.3.5"
+    id("com.expediagroup.graphql") version "7.0.1"
+}
+
+application {
+    mainClass.set("com.baeldung.graphql.server.KtorApplicationKt")
+}
+
+kotlin {
+    jvmToolchain(20)
```
theangrydev, line 39: Which JRE is necessary? I noticed the jvmToolchain(20), and I noticed I was only able to build successfully if I used exactly JRE 20, JRE 21 for example I had gradle errors with. If this is expected, let’s mention somewhere in the article about the exact runtime environment requirements.

kotlin-ktor/src/main/java/com/baeldung/graphql/client/GraphQLService.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.baeldung.graphql.client.generated.SaveOrCreateAttendeeMutation
+import com.baeldung.graphql.client.generated.SaveOrCreateConferenceMutation
+import com.baeldung.graphql.client.generated.conferencebyidquery.Conference
+import com.baeldung.graphql.client.generated.objectbyidquery.ObjectWithId
+import com.baeldung.graphql.client.generated.saveorcreateattendeemutation.Attendee
+import com.expediagroup.graphql.client.jackson.GraphQLClientJacksonSerializer
+import com.expediagroup.graphql.client.ktor.GraphQLKtorClient
+import java.net.URI
+
+class GraphQLService(url: String) {
```
theangrydev, line 14: For clarity, can we rename this to `GraphQLClient` or similar, to emphasise it contains the client side code. This would then match the corresponding client test name.

kotlin-ktor/src/main/java/com/baeldung/graphql/server/KtorApplication.kt
```diff
@@ -0,0 +1,44 @@
...
+        pingPeriod = Duration.ofSeconds(1)
+        contentConverter = JacksonWebsocketContentConverter()
+    }
+    install(GraphQL) {
+        schema {
+            packages = listOf("com.baeldung.graphql.server")
+            queries = listOf(ConferenceQuery())
+            mutations = listOf(ConferenceMutation(), AttendeeMutation())
+            subscriptions = listOf(ConferenceSubscription())
+        }
```
theangrydev, line 38: Out of curiosity, is there any support for classpath scanning of these instead of having to manually list them?

kotlin-ktor/src/main/java/com/baeldung/graphql/server/api/ConferencePublisher.kt
```diff
@@ -0,0 +1,17 @@
+package com.baeldung.graphql.server.api
+
+import com.baeldung.graphql.server.data.Conference
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+
+object ConferencePublisher {
+
+    private val mutableConferenceFlow: MutableSharedFlow<Conference> = MutableSharedFlow(replay = 1)
```
theangrydev, line 9: What is the significance of the `replay = 1` parameter?

kotlin-ktor/src/main/java/com/baeldung/graphql/server/api/ConferenceSubscription.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.expediagroup.graphql.generator.annotations.GraphQLDescription
+import com.expediagroup.graphql.server.operations.Subscription
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+class ConferenceSubscription : Subscription {
+
+    @GraphQLDescription("Emits single, newly created conferences")
+    fun conference(): Flow<Int> = ConferencePublisher.conferenceFlow.map { it.id!! }
```
theangrydev, line 11: Should we rename this to `conferenceIds` instead, to match the actual behaviour?

kotlin-ktor/src/main/resources/schema.graphql
theangrydev, line 1: See point on Jira about where this comes from

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.BeforeClass
+import kotlin.test.*
+
+class GraphQLClientTest {
+
+    private val graphQLService = GraphQLService("http://0.0.0.0:8080/graphql")
+
+    // https://stackoverflow.com/questions/61245981/how-can-i-start-ktor-server-programmatically-for-integration-tests
+    // Since regular Ktor tests work with a dummy application with no server, we can't actually test the client properly without starting it
+    // So we start it up and tear it down for every test.
```
theangrydev, line 19: Do you think this is something particularly confusing (and so worth mentioning) or something that Ktor users would already be familiar with as a test design pattern?

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        val fetchedConference = graphQLService.findConferenceById(0)
+        assertNull(fetchedConference)
+    }
+
+    @Test
+    fun `given create or save conference when saving not existing conference then should create it`(): Unit = runBlocking {
+        graphQLService.createOrSaveConference(null, "Some Conference")
+        val createdConference = ConferenceRepository.findById(0)
+        assertEquals("Some Conference", createdConference?.name)
+    }
```
theangrydev, line 80: What about the freshly assigned conference id? Is it correctly returned in the response to the client?

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        val updatedConference = ConferenceRepository.findById(0)
+        assertEquals("Same Conference", updatedConference?.name)
+    }
+
+    @Test
+    fun `given create or save attendee when saving not existing attendee then should create it`(): Unit = runBlocking {
+        graphQLService.createOrSaveAttendee(null, "Some Attendee")
+        val createdAttendee = AttendeeRepository.findById(0)
+        assertEquals("Some Attendee", createdAttendee?.name)
+    }
```
theangrydev, line 95: What about the freshly assigned attendee id? Is it correctly returned in the response to the client?

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        createDummyConference()
+        val objectById = graphQLService.getObjectById(0)
+        assertIs<com.baeldung.graphql.client.generated.objectbyidquery.Conference>(objectById)
+    }
+
+    @Test
+    fun `given get object by id when fetching attendee then should return it`(): Unit = runBlocking {
+        createDummyAttendee()
+        val objectById = graphQLService.getObjectById(0)
+        assertIs<com.baeldung.graphql.client.generated.objectbyidquery.Attendee2>(objectById)
```
theangrydev, line 116: Interesting, this `Attendee2` highlights even more this issue of generating many types with the same name that vary slightly depending on the area they were generated from. I'm surprised this issue/limitation hasn't received more attention so far. No action here, just commenting.

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    fun `given get object by id when fetching not existing object then should return null`(): Unit = runBlocking {
+        assertNull(graphQLService.getObjectById(0))
+    }
+
+    @Test
+    fun `given conference with attendees when fetching conference with limit one then should return conference with one attendee object`(): Unit = runBlocking {
+        createDummyAttendee()
+        createDummyAttendee()
+        createDummyConference(listOf(0, 1))
+        val fetchedConference = graphQLService.findConferenceById(0, 1)
```
theangrydev, line 129: For clarity, can we use named parameters:
```suggestion
        val fetchedConference = graphQLService.findConferenceById(id = 0, attendeeLimit = 1)
```

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        assertEquals(1, fetchedConference?.attendeeObjects?.size)
+    }
+
+    @Test
+    fun `given conference batch when fetching two conferences then should return both`() = runBlocking {
+        createDummyConference()
+        createDummyConference()
+        val conferences = graphQLService.getConferenceBatch(0, 1)
```
theangrydev, line 137: For clarity, can we use named parameters:
```suggestion
        val conferences = graphQLService.getConferenceBatch(firstId = 0, secondId = 1)
```

kotlin-ktor/src/test/java/com/baeldung/graphql/server/AttendeeTest.kt
```diff
@@ -0,0 +1,76 @@
...
+private const val UPDATE_ATTENDEE_QUERY = """
+    mutation {
+        saveOrCreateAttendee(attendee: { id: 0, name: "Jake Jakeson" }) {
+            id
+            name
+      }
+    }
+    """
+
+class AttendeeTest {
```
theangrydev, line 34: Can we rename this to `GraphQLServerAttendeeTest` or similar, to emphasise we are testing the server side in this test?

kotlin-ktor/src/test/java/com/baeldung/graphql/server/ConferenceTest.kt
```diff
@@ -0,0 +1,141 @@
...
+private const val FIND_BY_ID_QUERY = """
+    {
+        conferenceById(id: 0) {
+            name
+            id
+        }
+    }
+    """
+
+class ConferenceTest {
```
theangrydev, line 46: Can we rename this to `GraphQLServerConferenceTest` or similar, to emphasise we are testing the server side in this test?

theangrydev, line 46: Since all the other mutations/queries are covered, should we also include an `objectById` test here too for completeness?

kotlin-ktor/src/test/java/com/baeldung/graphql/server/AttendeeTest.kt
theangrydev: Since all the other mutations/queries are covered, should we also include an `objectById` test here too for completeness?

kotlin-ktor/src/test/java/com/baeldung/graphql/server/ConferenceTest.kt
```diff
@@ -0,0 +1,141 @@
...
+        with(call) {
+            assertEquals(HttpStatusCode.OK, status)
+            val response = jacksonObjectMapper().readValue(bodyAsText(), Map::class.java)
+            assertNotNull((response["data"] as Map<*, *>)["saveOrCreateConference"])
+            assertEquals("My Conference", ConferenceRepository.findById(0)?.name)
+        }
+    }
+
+    @Test
+    fun `when creating a conference then should trigger subscription`(): Unit = graphQlTestEnvironment { client ->
```
theangrydev, line 122: Let's mention WebSocket in the test name to highlight it is doing something quite different than the other tests

kotlin-ktor/src/test/java/com/baeldung/graphql/server/AttendeeTest.kt
theangrydev: Add the TestCase import

theangrydev (changes requested):
This is really great sample code, thanks a lot for taking the time to create this!

kotlin-ktor/src/main/java/com/baeldung/graphql/server/KtorApplication.kt
theangrydev: Also, according to https://opensource.expediagroup.com/graphql-kotlin/docs/schema-generator/schema-generator-getting-started/

I was expecting to see a toSchema call using the graphql-kotlin-schema-generator

theangrydev: Looks like the current PR actually does not use graphql-kotlin-schema-generator, if I comment out that dependency the build is still green

theangrydev: Have a look at the task `graphqlGenerateSDL` https://opensource.expediagroup.com/graphql-kotlin/docs/plugins/gradle-plugin-tasks

I added in `build.gradle.kts`:
```
graphql {
    schema {
        packages = listOf("com.baeldung.graphql.server")
    }
}
```

This currently fails with:
```
Execution failed for task ':graphqlGenerateSDL'.
> There was a failure while executing work items
   > A failure occurred while executing com.expediagroup.graphql.plugin.gradle.actions.GenerateSDLAction
      > Schema requires all subscriptions to be public and return a valid type from the hooks. ConferenceSubscription has PUBLIC visibility modifier. The function return type is Flow
```

I was not able to figure out how to resolve it, but after commenting out the `ConferenceSubscription` entirely, it did correctly generate `build/schema.graphql` for the remaining types.

If we can figure out how to resolve this failure for the subscription, we can then stop hard coding the `src/main/resources/schema.graphql` file and replace it with the generated one.

It should also be possible to order the `./gradlew build` tasks in such a way that `graphqlGenerateSDL` is called before `graphqlGenerateClient`.

theangrydev:
One more overall comment, can you optimise imports for the project, there were a few files with unused imports

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import com.baeldung.graphql.server.data.ConferenceRepository
+import com.baeldung.graphql.server.graphQLModule
+import io.ktor.server.engine.*
+import io.ktor.server.netty.*
+import kotlinx.coroutines.runBlocking
+import org.junit.AfterClass
+import org.junit.BeforeClass
+import kotlin.test.*
+
+class GraphQLClientTest {
```
theangrydev, line 13: In some of these tests, we only assert on the content of the repository, but do not check the result of the client call itself. This is particularly important in the case where we create with a null id, and then the client might rely on that id for future lookups.

For those cases, can we do something along the lines of:
```
        val response = graphQLService.createOrSaveAttendee(null, "Some Attendee")
        assertEquals(0, response?.id)
        assertEquals("Some Attendee", response?.name)

        val repository = AttendeeRepository.findById(0)
        assertEquals("Some Attendee", repository.name)
```

To make it clear that there is consistency between the repository and the response returned to the client.

kotlin-ktor/gradlew
VladGaluska: ✅ for gradle wrapper
For chmod, I'm not sure how to fix this, I can try running it on my other Mac PC if you think it makes sense

kotlin-ktor/build.gradle.kts
VladGaluska: ✅

kotlin-ktor/build.gradle.kts
VladGaluska: ✅ Answered on Jira (technical-1). JRE 20 is the latest supported JRE by Kotlin

kotlin-ktor/src/main/java/com/baeldung/graphql/client/GraphQLService.kt
VladGaluska: ✅ I thought this might cause confusion with the expedia group client, but they do have the added 'Ktor' in it, so I think it's alright

kotlin-ktor/src/main/java/com/baeldung/graphql/server/KtorApplication.kt
VladGaluska: ✅Left a discussion on Jira

kotlin-ktor/src/main/java/com/baeldung/graphql/server/api/ConferencePublisher.kt
VladGaluska: When a new collector subscribes to the flow, he will be emitted the last `replay` values emitted. I would leave a comment but I think this is described well in the `MutableSharedFlow` docs which are easily accessible from IDE's by navigating to them

kotlin-ktor/src/main/java/com/baeldung/graphql/server/api/ConferenceSubscription.kt
VladGaluska: Renamed to `conferenceId`

kotlin-ktor/src/main/resources/schema.graphql
VladGaluska: Done

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: The usual behavior for these tests is to use a dummy server, and it's a bit obscure to find data on setting up an application from scratch for tests

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: ✅ Good point, missed that test case

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: ✅

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: ✅

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: ✅

kotlin-ktor/src/test/java/com/baeldung/graphql/server/AttendeeTest.kt
VladGaluska: 1. Since this is in the server package, I'm not sure this really makes sense?
2. ✅
3. Not sure what you mean, the import is there for me?

kotlin-ktor/src/test/java/com/baeldung/graphql/server/ConferenceTest.kt
VladGaluska: Since this is in the server package, I'm not sure this really makes sense?

VladGaluska: ✅

kotlin-ktor/src/test/java/com/baeldung/graphql/server/ConferenceTest.kt
VladGaluska: ✅

kotlin-ktor/src/test/java/com/baeldung/graphql/client/GraphQLClientTest.kt
VladGaluska: The thing is, I wanted to avoid this for all the tests as it really clutters the readability of the tests and we start testing out-of-scope behavior in tests meant to check just one part of the functionality. I think this makes sense with the tests that cover creation, but not so sure about the other tests. What do you think?

theangrydev: Yes agreed as long as it is covered once somewhere, checks elsewhere can be looser

kotlin-ktor/src/test/java/com/baeldung/graphql/server/AttendeeTest.kt
theangrydev: 1. I think I meant to be consistent with the `GraphQLClientTest` naming, which is in the client package
3. I meant, add the static import so that `TestCase.assertEquals` becomes just `assertEquals`

kotlin-ktor/src/test/java/com/baeldung/graphql/server/ConferenceTest.kt
theangrydev: I think I meant to be consistent with the `GraphQLClientTest` naming, which is in the client package

theangrydev (approved):
Responded to the outstanding comments here, I'm okay to approve the PR as-is but feel free to make further changes if you feel like it

PR #668, opened 2023-10-26
JAVA-9433 - Using Ktlint to check TestNaming convention

GaetanoPiazzolla:
I've used an ant-task as the available maven plugins do not work. 
The rule will check if the test ends in the correct way. 
right now is disabled.


ktlint-custom/src/main/kotlin/com/baeldung/ktlintcustomrule/TestNamingRule.kt
```diff
@@ -0,0 +1,48 @@
+package com.baeldung.ktlintcustomrule
+
+import com.pinterest.ktlint.rule.engine.core.api.ElementType
+import com.pinterest.ktlint.rule.engine.core.api.Rule
+import com.pinterest.ktlint.rule.engine.core.api.RuleId
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+
+class TestNamingRule :
+    Rule(
+        ruleId = RuleId("$CUSTOM_RULE_SET_ID:unit-test-convention"),
```
anastasiosioannidis, line 10: about = About(maintainer = "Gaetano Piazzolla"),
  --> could we remove this section, or is it required?

PR #675, opened 2023-11-01
when_comparisons

kanake10:


core-kotlin-modules/core-kotlin-lang-4/src/test/kotlin/com/baeldung/pair/when/EquaToTest.kt
```diff
@@ -0,0 +1,22 @@
+package com.baeldung.pair.`when`
```
Bwvolleyball, line 1: I noticed that this package is using `when` as a package name, since that's a keyword in Kotlin, and this article is an advanced usage of `when()`, let's change that package to `whencomparisons` or something like that.

kanake10: On it !!

PR #678, opened 2023-11-05
KTLN-705: Removing an Entry From a Map in Kotlin While Iterating Over It

tapankavasthi:
## Summary
Removing an Entry From a Map in Kotlin While Iterating Over It

## Jira
[KTLN-705](https://jira.baeldung.com/browse/KTLN-705)

### DIFF
--- a/core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removeentry/RemoveMapEntryDuringIterationUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removeentry/RemoveMapEntryDuringIterationUnitTest.kt
@@ -1,0 +1,62 @@
+package com.baeldung.removeentry
+
+import org.junit.jupiter.api.Test
+import java.util.ConcurrentModificationException
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RemoveMapEntryDuringIterationUnitTest {
+
+    @Test
+    fun `test removing entry from map while iterating using for loop should remove successfully`() {
+        val map = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+        val iterator = map.entries.iterator()
+
+        while (iterator.hasNext()) {
+            val entry = iterator.next()
+            if (entry.key % 2 == 0) {
+                iterator.remove()
+            }
+        }
+
+        assertEquals(4, map.size)
+    }
+
+
+    @Test
+    fun `test removing entry from map while iterating with for loop should fail with ConcurrentModificationException`() {
+        val map = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+        assertFailsWith<ConcurrentModificationException> {
+            for ((key, value) in map) {
+                if (key % 2 == 0) {
+                    map.remove(key)
+                }
+            }
+        }
+    }
+
+    @Test
+    fun `test removing entry from map using predicate in removeIf method should remove successfully`() {
+        val map = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+        map.entries.removeIf { it.key % 2 == 0 }
+        assertEquals(4, map.size)
+    }
+
+    @Test
+    fun `test removing entry from map using mark and remove strategy should remove successfully`() {
+        val map = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+
+        val candidates = HashSet<Int>()
+        // mark phase
+        for ((key, value) in map) {
+            if (key % 2 == 0) {
+                candidates.add(key)
+            }
+        }
+        // delete phase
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removekey/RemoveKeyDuringIterationUnitTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import org.junit.jupiter.api.Test
+import java.util.ConcurrentModificationException
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RemoveKeyDuringIterationUnitTest {
+
+    @Test
+    fun `test removing key from map while iterating using for loop should pass`() {
```
theangrydev, line 11: Instead of saying "should pass" let's specify what the expectation actually is, something like:
```suggestion
    fun `test removing key from map while iterating using for loop should remove successfully`() {
```

Similar for other tests.

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removekey/RemoveKeyDuringIterationUnitTest.kt
```diff
@@ -0,0 +1,63 @@
+package com.baeldung.removekey
+
+import org.junit.jupiter.api.Test
+import java.util.ConcurrentModificationException
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class RemoveKeyDuringIterationUnitTest {
```
theangrydev, line 8: Let's rename this to better reflect what is going on. We remove entries from the map using their keys, rather than removing keys.

Similar changes should be make to the package name, and the test names.

```suggestion
class RemoveMapEntryDuringIterationUnitTest {
```

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removekey/RemoveKeyDuringIterationUnitTest.kt
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+    }
+
+    @Test
+    fun `test removing key from map using predicate in removeIf method should pass`() {
+    val mutableMap = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+    val isEven: (Int) -> Boolean = { it -> it % 2 == 0 }
+    mutableMap.entries.removeIf { isEven(it.key) }
+    assertEquals(4, mutableMap.size)
```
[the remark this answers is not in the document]
tapankavasthi: It's fixed now.

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removekey/RemoveKeyDuringIterationUnitTest.kt
tapankavasthi: Thank you for suggesting this change. I've incorporated these and revised the PR.

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removekey/RemoveKeyDuringIterationUnitTest.kt
tapankavasthi: Perfect! The new class name makes more sense now. Thanks for suggesting!

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removeentry/RemoveMapEntryDuringIterationUnitTest.kt
```diff
@@ -0,0 +1,63 @@
...
+                }
+            }
+        }
+    }
+
+    @Test
+    fun `test removing entry from map using predicate in removeIf method should remove successfully`() {
+        val mutableMap = mutableMapOf(1 to 3, 2 to 3, 3 to 2, 4 to 2, 7 to 2, 9 to 1)
+        val isEven: (Int) -> Boolean = { it -> it % 2 == 0 }
+        mutableMap.entries.removeIf { isEven(it.key) }
```
theangrydev, line 42: I think the predicate example might be simpler to understand with one less layer of indirection:
```suggestion
    mutableMap.entries.removeIf { it.key % 2 == 0 }
```

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/removeentry/RemoveMapEntryDuringIterationUnitTest.kt
tapankavasthi: Done

theangrydev (approved):

PR #682, opened 2023-11-08
KTL-679: Create a CRUD service with Kotlin and Spring Boot

harry9656:


spring-boot-crud-kotlin/src/test/kotlin/com/baeldung/controllers/TaskCRUDIntegrationTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    var taskId: Long = 0
+
+    @Test
+    fun createTask() {
+        val taskData = TaskData("Task", "description", false)
+        val result = this.restTemplate.postForEntity("/tasks/create", taskData, TaskEntity::class.java)
+        taskId = result?.body?.id!!
+        assertTrue { result?.body?.name.equals("Task") }
+        assertTrue { result?.body?.description.equals("description") }
```
JavierPeris, line 29: I think the safe call on result is not needed so you can, do the dot call directly:

```suggestion
        assertTrue { result.body?.name.equals("Task") }
        assertTrue { result.body?.description.equals("description") }
```

PR #684, opened 2023-11-09
It and this

kanake10:


core-kotlin-modules/core-kotlin-lang-4/src/test/kotlin/com/baeldung/pair/itandthiskeyword/UsingThisTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class UsingThisTest {
+
+    data class Person(val name: String) {
+        // An instance variable
+        var age: Int = 0
+        fun printDetails() {
+            println("Name: $name")
+            println("Age: $age")
+        }
```
Bwvolleyball, line 14: Since we no longer reference this in the article, let's delete this.

PR #705, opened 2023-11-27
[KTLN-768] Add Samples

LeoColman:


core-kotlin-modules/core-kotlin-9/src/test/kotlin/com/baeldung/ifvslet/IfVsLetUnitTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        val uppercased = if(name != null) name.uppercase() else null
+
+        if (name == null) {
+            assertNull(uppercased)
+        } else {
+            assertEquals("NAME", uppercased)
+        }
+    }
+
+    fun fetchNameFromDatabase(): String? = if(Random.nextBoolean()) "name" else null
```
Bwvolleyball, line 34: Let's re-work this to not have randomized tests. Can we just split this into 4 tests - one each for null and non-null?

Bwvolleyball (commented):
Just a couple formatting issues that I should be able to fix via the comments.

PR #707, opened 2023-11-27
KTLN-786: Reverse the order of words in a sentence in Kotlin

nappy29:
Added unit test


core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/sentencereverse/ReverseSentenceUnitTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    fun `reverse sentence using a stack`() {
+        val sentence = "this is a complete sentence"
+        assertEquals("sentence complete a is this", reverseWordsInSentenceUsingStack(sentence))
+    }
+
+    @Test
+    fun testReverseWordsInSentence() {
+        assertEquals("this is a complete sentence", reverseWordsInSentence("sentence complete a is this"))
+    }
```
Bwvolleyball, line 36: It looks like this example didn't actually make it to the article, so let's take it out.

core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/sentencereverse/ReverseSentenceUnitTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    fun reverseSentenceUsingReverseAndJoinToStringMethods(sentence: String): String {
+        val words = sentence.split(" ")
+        return words.reversed().joinToString(" ")
+    }
+
+    fun reverseWordsInSentence(sentence: String): String {
+        val words = sentence.split(" ")
+        return words.reduce { acc, s -> "$s $acc" }
+    }
```
Bwvolleyball, line 46: This also didn't make it to the draft, so we can take it out.

PR #711, opened 2023-11-28
JAVA-28626 | ktlint library committed in repository.

GaetanoPiazzolla:


pom.xml
```diff
@@ -322,6 +368,60 @@
                 <version>${gitflow-incremental-builder.version}</version>
             </extension>
         </extensions>
+
```
anastasiosioannidis, line 371: Just a question, why do we need to ignore below executions of 2 plugins for Eclipse?

PR #752, opened 2023-12-31
underscore

kanake10:


core-kotlin-modules/core-kotlin-9/src/test/kotlin/com/baeldung/underscoreoperator/UnderScoreOperator.kt
```diff
@@ -0,0 +1,34 @@
...
+inline fun <T> printElementInfo(element: T) {
+}
+fun mainForHOF() {
+    printElementInfo<_>(42)
+    printElementInfo<_>("Hello")
+}
+
+
+
+
```
Bwvolleyball, line 34: Extra spaces we should get rid of.
```suggestion

```

core-kotlin-modules/core-kotlin-9/src/test/kotlin/com/baeldung/underscoreoperator/UnderScoreOperator.kt
```diff
@@ -0,0 +1,34 @@
...
+fun mainTwo(args: Array<String>) {
+    val list = listOf("hi")
+    val list2 = list.mapIndexed { _, item -> item }
+    println(list2)
+}
+
+
+inline fun <T> printElementInfo(element: T) {
+}
```
Bwvolleyball, line 26: Can we still just print the element?

```suggestion
inline fun <T> printElementInfo(element: T) {
    println("Element: $element")
}
```

PR #767, opened 2024-01-08
enums

kanake10:


core-kotlin-modules/core-kotlin-9/src/test/kotlin/com/baeldung/enum/Enum.kt
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.enum
```
Bwvolleyball, line 1: Just noticed this - but this package name is a bit weird.

Can we change the package to

```suggestion
package com.baeldung.whenenum
```

or something like that instead?

PR #786, opened 2024-01-23
collect

kanake10:


core-kotlin-modules/core-kotlin-concurrency-3/src/main/java/com/baeldung/comparecollectandcollectlatest/UsingCollect.kt
```diff
@@ -0,0 +1,13 @@
...
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.runBlocking
+
+fun main() {
+    runBlocking {
+        flowOf(1, 2, 3)
+                .collect { value ->
+                    println(value)
+                }
+    }
```
Bwvolleyball, line 12: This formatting feels a bit off - would it make sense to arrange it like this instead?
The change is to put `.collect()` starting on the same line of flowOf(), so that we don't need quite so many indents.

This will also apply to the other snippet.

```kotlin
        flowOf(1, 2, 3).collect { value ->
            println(value)
        }
    }
```

core-kotlin-modules/core-kotlin-concurrency-3/src/main/java/com/baeldung/comparecollectandcollectlatest/UsingCollect.kt
```diff
@@ -0,0 +1,12 @@
...
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.runBlocking
+
+fun main() {
+    runBlocking {
+        flowOf(1, 2, 3).collect { value ->
+                    println(value)
+        }
+    }
+}
```
Bwvolleyball, line 12: Should this `println()` be as indented as it is, or should it be less?

```suggestion
fun main() {
    runBlocking {
        flowOf(1, 2, 3).collect { value ->
            println(value)
        }
    }
}
```

core-kotlin-modules/core-kotlin-concurrency-3/src/main/java/com/baeldung/comparecollectandcollectlatest/UsingCollectLatest.kt
```diff
@@ -0,0 +1,14 @@
...
+import kotlinx.coroutines.runBlocking
+fun main() {
+    runBlocking {
+        flowOf(1, 2, 3).collectLatest { value ->
+                    println(value)
+        }
+    }
+}
+
+
```
Bwvolleyball, line 14: Same with the spacing here.

```suggestion
import kotlinx.coroutines.runBlocking

fun main() {
    runBlocking {
        flowOf(1, 2, 3).collectLatest { value ->
            println(value)
        }
    }
}

```

PR #795, opened 2024-02-03
KTLN-452: Creating a Map from Two Arrays in Kotlin

nappy29:
Added Unit tests


core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/mapFromTwoArrays/MapFromTwoArraysUnitTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val map = keys.zip(values).toMap()
+        val expected = mapOf("a" to 1, "b" to 2, "c" to 3)
+
+        assertEquals(3, map.size)
+        assertEquals(expected, map)
+    }
+
+    @Test
+    fun `creates a map from two arrays using the associate() method`() {
```
Bwvolleyball, line 33:
```suggestion
    fun `creates a map from two arrays using the associateWith() method`() {
```

PR #806, opened 2024-02-12
[sb-@Value] @Value inject in Kotlin

sk1418:


spring-boot-kotlin/src/main/kotlin/com/baeldung/theValueAnnotation/MyMagicBeans.kt
```diff
@@ -0,0 +1,27 @@
+package com.baeldung.theValueAnnotation
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+@Component
+class MyMagicBean(
```
theangrydev, line 7: Can we rename this to something more descriptive of its purpose?

Maybe something like PropertyValues or PropertyValueExamples or ValueExamples or similar, your choice, just MyMagicBean struck me as a bit too cryptic

Alternatively, happy to be convinced otherwise if it was supposed to be a fun example

sk1418: I wanted to pick an interesting name. I agree with you after reading you r comment.
I renamed it to `ValueBean` I hope it's ok, as I want readers to realize that this is a Spring Bean/Component.

PR #815, opened 2024-02-19
spies

kanake10:


kotlin-testing/src/test/kotlin/com/baeldung/usingspy/CalculatorTests.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class Calculator{
+    fun add(a: Int, b: Int): Int {
+        return a+b
+    }
+    fun findAverage(a: Int, b: Int): Int {
+        val total = add(a,b)
+        return total/2;
+    }
+}
```
Bwvolleyball, line 18: Let's add a little spacing in here:
```suggestion
class Calculator {
    fun add(a: Int, b: Int): Int {
        return a + b
    }
    fun findAverage(a: Int, b: Int): Int {
        val total = add(a, b)
        return total / 2;
    }
}
```

kotlin-testing/src/test/kotlin/com/baeldung/usingspy/CalculatorTests.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class CalculatorTest {
+    @Test
+    fun testSpy() {
+        val spy = spyk<Calculator>()
+        val result = spy.findAverage(5,5)
```
Bwvolleyball, line 24: And here:
```suggestion
        val result = spy.findAverage(5, 5)
```

kotlin-testing/src/test/kotlin/com/baeldung/usingspy/CalculatorTests.kt
Bwvolleyball, line 1: The build failed with this error:

```
[INFO]    [ktlint] /home/jenkins/workspace/kotlin-tutorials-build-pr/kotlin-testing/src/test/kotlin/com/baeldung/usingspy/CalculatorTests.kt:20:7: Unit test class names need to end in 'UnitTest', integration tests with 'IntegrationTest', etc (custom-rule-set-id:unit-test-convention)
```

But let's rename this file to be `SpiesUnitTest`, since the point of this isn't to test the calculator, but rather the spies.

PR #817, opened 2024-02-21
[KTLN-501] Accessing Methods Outside Companion Object in Kotlin

albertache1998:
code refactor


core-kotlin-modules/core-kotlin-10/src/test/kotlin/com/baeldung/methodAccessInCompanionObject/MethodAccessInCompanionObjectUnitTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    fun outerClassMethod(): String {
+        return "This is a method outside the companion object"
+    }
+}
+interface OuterClassInterface {
+    fun outerClassMethod(): String
+}
+
+class OuterClassWIthInterface : OuterClassInterface {
```
KevinGilmore, line 54: There's an uppercase 'I' in the "With" part of the name. Lowercase it so the name is OuterClassWithInterface.

PR #820, opened 2024-02-23
KTLN-522: In-Place Modification of Map Entry in Kotlin

nappy29:
unit test code

### DIFF
--- a/core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/modifyMapEntryInPlace/ModifyMapEntryInPlaceUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/modifyMapEntryInPlace/ModifyMapEntryInPlaceUnitTest.kt
@@ -13,7 +13,7 @@ class ModifyMapEntryInPlaceUnitTest {
     }
 
     @Test
-    fun `modifies mutable map entry using assign operator`() {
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-collections-map/src/test/kotlin/com/baeldung/modifyMapEntryInPlace/ModifyMapEntryInPlaceUnitTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    @Test
+    fun `modifies mutable map entry using put method`() {
+        val map = mutableMapOf("key1" to "value1", "key2" to "value2")
+        map.put("key1", "new value")
+        assertEquals("new value", map["key1"])
+    }
+
+    @Test
+    fun `modifies mutable map entry using assign operator`() {
```
Bwvolleyball, line 16: might need to update the name of this test with the name of the section.

Bwvolleyball: To match the verbiage in the draft, let's make this `set operator` instead - using pr comments to be able to update this.

```suggestion
    fun `modifies mutable map entry using set operator`() {
```

PR #823, opened 2024-02-26
KTLN-351 | Cloning an object in Kotlin | Hangga Aji Sayekti

hangga:


core-kotlin-modules/core-kotlin-10/src/test/kotlin/com/baeldung/cloningobject/CloningObjectUnitTest.kt
```diff
@@ -0,0 +1,74 @@
...
+class CloningObjectUnitTest {
+
+    private val address = Address("Jln. Kemasan", "Yogyakarta")
+    private val person = Person("Hangga", address)
+
+    @Test
+    fun `when using secondary constructor then proves that shallow copy`() {
+        val clonedPerson = Person(person)
+        clonedPerson.address.city = "Surabaya"
+        clonedPerson.address.street = "Jln. Ahmad Yani"
```
theangrydev, line 29: I think it would be clearer if we modified the original person rather than the cloned person in these tests, and showed the impact that does or doesn't have.

I also think it would be clearer if we did the isEqualTo using hard coded string constants rather than field references. This would also remove the need for the println statements, since we would be able to see the expected strings already in the test.

Let me know what you think - let's do a round on just the PR first before going back to the article, to avoid rework.

hangga: Oh yeah, I agree.
that makes more sense.
I have done it just now. Please review again.
next, I'll update the article.

core-kotlin-modules/core-kotlin-10/src/test/kotlin/com/baeldung/cloningobject/CloningObjectUnitTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    fun `when own function then proves that deep copy`() {
+        val clonedPerson = person.deepCopy()
+        assertThat(clonedPerson).isNotSameAs(person)
+
+        person.address.city = "Bandung"
+        person.address.street = "Jln. Siliwangi"
+
+        assertThat(clonedPerson.address.city).isNotEqualTo("Bandung")
+        assertThat(clonedPerson.address.street).isNotEqualTo("Jln. Siliwangi")
```
theangrydev, line 70: I almost missed the isNotEqualTo when reading through this - how about we reformat each test assertions so we can see the isEqualTo/isNotEqualTo more easily?
```suggestion
        assertThat(clonedPerson.address.city)
            .isNotEqualTo("Bandung")
        assertThat(clonedPerson.address.street)
            .isNotEqualTo("Jln. Siliwangi")
```

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-10/src/test/kotlin/com/baeldung/cloningobject/CloningObjectUnitTest.kt
hangga: Oh, yeah right.
I will implement your suggestion.

theangrydev (changes requested):
Let's add one more test to show that the basic copy() is a shallow copy. This doesn't need to be mentioned in the article, just something to add to the PR.

theangrydev (approved):

hangga:
Hey @theangrydev it's still open

PR #825, opened 2024-02-27
[KTLN-827] Java Boolean.valueOf() Equivalent in Kotlin

albertache1998:
Added unit test

### DIFF
--- a/core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/booleanValueOfKotlinEquivalent/BooleanValueOfKotlinEquivalentUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/booleanValueOfKotlinEquivalent/BooleanValueOfKotlinEquivalentUnitTest.kt
@@ -1,6 +1,7 @@
 package com.baeldung.booleanValueOfKotlinEquivalent
 
 import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertThrows
 import org.junit.jupiter.api.Test
 
 class BooleanValueOfKotlinEquivalentUnitTest {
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/booleanValueOfKotlinEquivalent/BooleanValueOfKotlinEquivalentUnitTest.kt
```diff
@@ -0,0 +1,83 @@
+package com.baeldung.booleanValueOfKotlinEquivalent
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+
+class BooleanValueOfKotlinEquivalentUnitTest {
+
+    @Test
+    fun `obtain equivalent in Kotlin using if statement`() {
```
Bwvolleyball, line 10: Let's rename this test to be around string comparison instead.

core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/booleanValueOfKotlinEquivalent/BooleanValueOfKotlinEquivalentUnitTest.kt
```diff
@@ -0,0 +1,83 @@
...
+fun convertStringToBooleanUsingRegex(input: String): Boolean {
+    return input.matches("^\\s*(?i:true)\\s*$".toRegex())
+}
+
+fun String.toBooleanValue(): Boolean? =
+    when (this.toLowerCase()) {
+        "true" -> true
+        "false" -> false
+        else -> null
+    }
```
Bwvolleyball, line 83: This doesn't match the snippet from the article:

```suggestion
fun String.toBooleanValue(): Boolean =
    when (this.toLowerCase()) {
        "true" -> true
        else -> false
    }
```

core-kotlin-modules/core-kotlin-strings-4/src/test/kotlin/com/baeldung/booleanValueOfKotlinEquivalent/BooleanValueOfKotlinEquivalentUnitTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        assertEquals(false, convertStringToBooleanUsingRegex("false"))
+    }
+
+}
+fun convertStringToBoolean(input: String): Boolean {
+    return if (input.equals("true", ignoreCase = true)) true else false
+}
+
+fun convertStringToBooleanUsingRegex(input: String): Boolean {
+    return input.trim().matches("^\\s*(?i:true)\\s*$".toRegex())
```
Bwvolleyball, line 74: Because we trimmed the string, we don't need the regex to include whitespace.

```suggestion
    return input.trim().matches("^(?i:true)$".toRegex())
```

PR #832, opened 2024-03-05
BAEL-102744 Review remaining warnings in kotlin-build job

timis1:


kotlin-libraries/src/main/kotlin/com/baeldung/kotlin/arrow/FunctionalErrorHandlingWithEither.kt
```diff
@@ -1,9 +1,6 @@
 package com.baeldung.kotlin.arrow
 
-import arrow.core.Either
-import arrow.core.filterOrElse
-import arrow.core.flatMap
-import arrow.core.right
+import arrow.core.*
```
alvin577austria, line 3: Can you bring back the multiple imports please.

alvin577austria (changes requested):

PR #836, opened 2024-03-11
KTLN-426: Update to jakarta.jakartaee-api dependency.

harry9656:


jee-kotlin/src/main/kotlin/com/baeldung/jeekotlin/entity/Student.kt
```diff
@@ -1,6 +1,6 @@
 package com.baeldung.jeekotlin.entity
 
-import com.fasterxml.jackson.annotation.JsonProperty
+import jakarta.persistence.*
 import javax.persistence.*
```
JavierPeris, line 4: Is this import used?

JavierPeris (commented):
Shouldn't we use jakarta.ws.rs imports for ApplicationConfig and StudentResource classes?

jee-kotlin/src/main/kotlin/com/baeldung/jeekotlin/entity/Student.kt
harry9656: It's not necessary. I removed it.

PR #837, opened 2024-03-11
error handling

kanake10:


core-kotlin-modules/core-kotlin-concurrency-3/src/main/java/com/baeldung/errorhandling/ExceptionHandlingMethods.kt
```diff
@@ -0,0 +1,60 @@
...
+
+fun main2() = runBlocking {
+    val exceptionHandler = CoroutineExceptionHandler { _, exception ->
+        println("Caught an exception: ${exception.message}")
+    }
+    supervisorScope {
+        val job = launch(exceptionHandler) {
+            launch { delay(200); throw CustomException("An exception occurred!") }
+            launch { delay(100); println("This coroutine completes successfully.") }
+        }
```
Bwvolleyball, line 45: Apologies if I missed this before, but let's make these multi-lines.

```suggestion
        val job = launch(exceptionHandler) {
            launch {
                delay(200)
                throw CustomException("An exception occurred!")
            }
            launch {
                delay(100)
                println("This coroutine completes successfully.")
            }
        }
```

PR #852, opened 2024-03-23
KTLN-394 - Qualified this and implicit this in Kotlin

hangga:


core-kotlin-modules/core-kotlin-lang-oop-3/src/test/kotlin/com/baeldung/implicitAndQualifiedthis/ImplicitAndQualifiedthisUnitTest.kt
```diff
@@ -0,0 +1,130 @@
...
+            val a = this
+
+            fun printLine() = "Member function"
+
+            fun checkThis() {
+                val b = this // Implicit: refers to the current instance of Outer
+                assertEquals(Outer::class.java.name, b::class.java.name)
+
+                assertEquals(1, this.x + 1) // Implicit: refers to the x property of the current Outer instance
+                assertEquals(6, x + 1) // Implicit: refers to the x property of the current Outer instance
```
theangrydev, line 27: Can we just use x instead of x+1 here?

theangrydev (changes requested):
We could potentially split this up into separate tests to match the flow of the article, but I'm also okay to leave it as-is

theangrydev (approved):

PR #855, opened 2024-03-25
[mockk-matching-varargs] mockk vararg machters

sk1418:


kotlin-testing-2/src/test/kotlin/com/baeldung/mockk/MatchVarargsUnitTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    @Test
+    fun `when calling concatStringsBySpace() on the real object then get expected result`() {
+        val result = MyClass().joinBySpace("a", "b", "c", "d", "e")
+        assertEquals("a b c d e", result)
+    }
+
+    @Test
+    fun `when using anyVararg() on the mocked object then get expected result`() {
+        every { mockkObj.joinBySpace("a", "b", "c", *anyVararg(), "x", "y", "z") } returns "Wow, Kotlin rocks!"
```
theangrydev, line 28: Is it ever possible for anyVararg to not match? I guess not from what I see in the test, as we see even in the "zero arg" case it matches

kotlin-testing-2/src/test/kotlin/com/baeldung/mockk/MatchVarargsUnitTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        assertEquals("Wow, Kotlin rocks!", result)
+
+        val result2 = mockkObj.joinBySpace("a", "b", "c", "x", "y", "z")
+        assertEquals("Wow, Kotlin rocks!", result2)
+    }
+
+
+    @Test
+    fun `when using varargAll() on the mocked object then get expected result`() {
+        every { mockkObj.joinBySpace("a", "b", "c", *varargAll { it.startsWith("d") }, "z") } returns "Wow, Kotlin rocks!"
```
theangrydev, line 40: What happens in the "zero arg" case for varaargAll?

kotlin-testing-2/src/test/kotlin/com/baeldung/mockk/MatchVarargsUnitTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        assertThrows<MockKException> {
+            mockkObj.joinBySpace("a", "b", "c", "d1", "Baeldung", "z")
+        }.also { exception ->
+            assertTrue(exception.message!!.startsWith("no answer found"))
+        }
+    }
+
+    @Test
+    fun `when using varargAny() on the mocked object then get expected result`() {
+        every { mockkObj.joinBySpace("a", "b", "c", *varargAny { it.startsWith("d") }, "z") } returns "Wow, Kotlin rocks!"
```
theangrydev, line 54: What happens in the "zero arg" case for varargAny?

kotlin-testing-2/src/test/kotlin/com/baeldung/mockk/MatchVarargsUnitTest.kt
sk1418: it always matches. (any number of elements + any value).

kotlin-testing-2/src/test/kotlin/com/baeldung/mockk/MatchVarargsUnitTest.kt
sk1418: addressed

PR #866, opened 2024-04-03
[KTLN-799] Getting All Field Names of a Class Using Kotlin Reflection

albertache1998:
Added unit test


core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/referenceLastElementDuringIteration/ReferenceLastElementDuringIterationUnitTest.kt
```diff
@@ -0,0 +1,60 @@
+package com.baeldung.referenceLastElementDuringIteration
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class ReferenceLastElementDuringIterationUnitTest {
```
Bwvolleyball, line 6: I think this is extra / from another ticket.

PR #872, opened 2024-04-07
[KTLN-676] Add samples

LeoColman:


core-kotlin-modules/core-kotlin-10/src/test/kotlin/com/baeldung/dijkstra/DijkstraUnitTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class DijkstraUnitTest {
+
+    @Test
+    fun `Should calculate shortest path when using Dijkstra algorithm`() {
+        val graph = mapOf(
+            1 to listOf(
```
Maiklins, line 11: I think it's better to write the node number and list of edges on a single line (like in the article). That's much easier to read.

PR #930, opened 2024-05-16
[KTLN-848] Add samples

LeoColman:


core-kotlin-modules/core-kotlin-files/src/main/kotlin/com/baeldung/tempfile/TempFile.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlin.io.path.createTempFile
+import kotlin.io.path.writeText
+
+fun createTempFileWithKotlinExtensions(): Path {
+    val tempFile = createTempFile(prefix = "kotlinTemp", suffix = ".tmp")
+    println("Temporary file created with Kotlin extensions at: ${tempFile.toAbsolutePath()}")
+
+    tempFile.writeText("Kotlin Path Data")
+
+    return tempFile
```
Bwvolleyball, line 15: Let's also get the delete on exit line into the example, even if we don't show it all at once.

```suggestion
    tempFile.writeText("Kotlin Path Data")

    tempFile.toFile().deleteOnExit()

    return tempFile
```

PR #933, opened 2024-05-21
KTLN-447: Getting a Kotlin KClass from a Package Class Name String

nappy29:
Added unit tests


core-kotlin-modules/core-kotlin-advanced-3/src/test/kotlin/com/baeldung/obtainKclassFromPackageClassName/GetKclassUnitTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        .enableClassInfo()
+        .scan()
+        .getClassInfo(className)
+
+    if(classInfo != null){
+        return classInfo.loadClass().kotlin
+    }
+
+    return null
+}
```
Bwvolleyball, line 90: Since this is actually a platform type (`ClassInfo!`) that is nullable, let's just rework this to be directly nullable, and return with the null-safe operator.

```suggestion
    val classInfo: ClassInfo? = ClassGraph()
        .addClassLoader(ClassLoader.getSystemClassLoader())
        .enableClassInfo()
        .scan()
        .getClassInfo(className)

     return classInfo?.loadClass()?.kotlin  
}
```

PR #956, opened 2024-06-09
[enum-field-to-json] enum property in JSON

sk1418:


kotlin-json-2/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlinx</groupId>
+            <artifactId>kotlinx-serialization-core-jvm</artifactId>
+            <version>${serialization.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
```
theangrydev, line 48: Is this similar to what we do in other Kotlin modules? I'm surprised that some of this isn't already present in the kotlin-modules parent

kotlin-json-2/src/test/kotlin/com/baeldung/kotlin/serializeEnumField/EnumFieldKotlinxUnitTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.Json
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+
+class EnumFieldKotlinx1UnitTest {
```
theangrydev, line 17:
```suggestion
class EnumFieldKotlinxUnitTest {
```

theangrydev (changes requested):

kotlin-json-2/pom.xml
sk1418: cp/paste from the kotlin-json module, removed

theangrydev (approved):

PR #961, opened 2024-06-13
[sorted-functions] using sortWith in code

sk1418:


theangrydev (changes requested):
We should use SLF4J instead of println

theangrydev (approved):

PR #978, opened 2024-06-25
KTLN-877 - Update article "Parallel Operations on Kotlin Collections"

hangga:


core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -81,14 +77,12 @@ class ParallelOperationCollectionsUnitTest {
...
-                            person.setAdult()
-                            person
-                        }.await()
-                    )
+                    person.setAdult()
+                    Thread.sleep(1500)
+                    emit(person)
                 }
             }
+            .flowOn(Dispatchers.IO)
```
theangrydev, line 85: Can you help me understand the purpose of this? I had a look at https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flow-on.html

Should we put the flowOn call right after the flow call, rather than after the flatMapMerge? https://stackoverflow.com/questions/75884997/what-is-default-dispatcher-kotlin-flow-function

If not, what is the difference between those two approaches?

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -105,9 +99,12 @@ class ParallelOperationCollectionsUnitTest {
...
                 {
-                    Observable.just(it).subscribeOn(Schedulers.computation()).doOnNext { person ->
-                        person.setAdult()
-                    }
+                    Observable.just(it)
+                        .subscribeOn(Schedulers.computation())
+                        .doOnNext { person ->
+                            person.setAdult()
+                            Thread.sleep(1500)
+                        }
```
theangrydev, line 107: Besides the sleep call, is this just a reformatting?

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -128,9 +125,12 @@ class ParallelOperationCollectionsUnitTest {
...
```
theangrydev, line 133: Besides the sleep call, is this just a reformatting?

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -159,28 +160,5 @@ class ParallelOperationCollectionsUnitTest {
...
-            }
-            .filter { it.age > 15 }
-            .sortedBy { it.age }
-
-        executor.shutdown()
-
-        startTime.printTotalTime()
-
-        futures.assertOver15AndSortedByAge()
-    }
```
theangrydev, line 184: We could potentially show a solution where we collect a `List<Future<T>>` and transform that to a `List<T>`

Any reason for not showing this kind of solution?

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
hangga: Oh right, I've added it.
Because in this unit test we use _runBlocking_, so without _.flowOn(Dispatchers.IO)_, then I'm afraid flow will be run in the same context as the coroutine wrapper (in this case _runBlocking_)

theangrydev: Sorry I wasn't clear. My question is, what is the difference between solutions A and B below:

A:
```
 val filteredPeople = people.asFlow()
            .flatMapMerge { person ->
                flow {
                    emit(
                        async {
                            person.setAdult()
                            person
                        }.await()
                    )
                    person.setAdult()
                    emit(person)
                }
            }
            .flowOn(Dispatchers.IO)
            .filter { it.age > 15 }.toList()
            .sortedBy { it.age }
```

B:
```
 val filteredPeople = people.asFlow()
            .flatMapMerge { person ->
                flow {
                    emit(
                        async {
                            person.setAdult()
                            person
                        }.await()
                    )
                    person.setAdult()
                    emit(person)
                }.flowOn(Dispatchers.IO)
            }
            .filter { it.age > 15 }.toList()
            .sortedBy { it.age }
            
```

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -105,9 +99,12 @@ class ParallelOperationCollectionsUnitTest {
...
             .flatMap(
                 {
-                    Observable.just(it).subscribeOn(Schedulers.computation()).doOnNext { person ->
-                        person.setAdult()
-                    }
+                    Observable.just(it)
+                        .subscribeOn(Schedulers.computation())
+                        .doOnNext { person ->
+                            person.setAdult()
+                            Thread.sleep(1500)
```
theangrydev, line 106: Let's remove the sleep calls

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -128,9 +125,12 @@ class ParallelOperationCollectionsUnitTest {
...
```
theangrydev, line 132: Let's remove the sleep calls

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -59,7 +54,8 @@ class ParallelOperationCollectionsUnitTest {
 
         val filteredPeople = people
             .map { person ->
-                async {
+                async(Dispatchers.IO) {
+                    Thread.sleep(1500)
```
theangrydev, line 58: Let's remove the sleep calls

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
```diff
@@ -150,6 +150,7 @@ class ParallelOperationCollectionsUnitTest {
         val filteredPeople = people.parallelStream()
             .map { person ->
                 person.setAdult()
+                Thread.sleep(1500)
```
theangrydev, line 153: Let's remove the sleep calls

theangrydev (changes requested):

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
hangga: Oh, sorry, @theangrydev. I misunderstood your question.

As I understand it, code A uses the _Dispatchers.IO_ context for the entire flow, which includes _flatMapMerge_, filter, and subsequent operations. This is more efficient if all operations require execution in IO.

Code B only uses the _Dispatchers.IO_ context inside _flatMapMerge_, while _filter_ and subsequent operations will be executed in the default context or the context where the _Flow_ was originally created.

Since in unit tests we always use _runBlocking_ to run coroutines, I currently choose to use code A so that the entire flow (including filter and subsequent operations) is executed in IO.

hangga: But now, I also remove _async {}_ inside _emit_ because I think there is no need to create new coroutines inside flow anymore.

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
hangga: Yes, that's right. This is to make it easier to read.

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
hangga: What does it mean to be like this?

        val executor = Executors.newFixedThreadPool(people.size)
        val futures: List<Future<Person>> = people
            .map { person ->
                executor.submit(Callable {
                    person.setAdult()
                    person
                })
            }

        // List<Future<Person>> to List<Person>
        val results = futures
            .map { it.get() }
            .filter { it.age > 15 }
            .sortedBy { it.age }

        executor.shutdown()

        results.assertOver15AndSortedByAge()

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
theangrydev: @hangga 

In A, can/should we put the flowOn(Dispatchers.IO) call before the flatMapMerge?

It's not very intuitive that the entire flow uses Dispatchers.IO when the flowOn call is made after the flarMapMerge

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
theangrydev: @hangga yes that's what I had in mind. I've used an approach similar to this in the past in plain Java.

hangga: After you agreed, I am more confident now. Because before, I was not sure about what I was doing.

core-kotlin-modules/core-kotlin-collections-6/src/test/kotlin/com/baeldung/parallelOperationsCollections/ParallelOperationCollectionsUnitTest.kt
hangga: Oh yeah, you're right.
It makes more sense to call _flowOn(Dispatchers.IO)_ before the _flatMapMerge_.
Thanks.

theangrydev (approved):

PR #1024, opened 2024-07-16
JAVA-37523 Update README.md

lor6:


README.md
```diff
@@ -17,6 +17,9 @@ or if we want to build the entire repository with Integration Tests enabled, we
 
 This should not be needed often as we are usually concerned with a specific module.
 
+To build the modules that use Gradle, use the command:
```
yadavan88, line 20: To build all the gradle modules, use the profile `gradle-modules` with the command:
`mvn clean install -Pgradle-modules`

This compiles and runs the tests in all the gradle submodules using the maven-exec plugin. 

However, we can build the individual gradle sub-modules using the `gradlew` command:
`./gradlew build` or `./gradlew test`. Additionally, we can also use the IDE such as IntelliJ IDEA to build and test the individual modules.

PR #1034, opened 2024-07-23
[KTLN-828] Add samples

LeoColman:


kotlin-libraries-data/src/test/kotlin/com/baeldung/retrofit/EnumSerializationUnitTest.kt
```diff
@@ -0,0 +1,106 @@
...
+data class Order(val id: String, val status: OrderStatus)
+
+interface OrderService {
+
+    @GET("orders/{id}")
+    suspend fun getOrder(@Path("id") orderId: String): Order
+
+    @POST("orders")
+    suspend fun createOrder(@Body order: Order): Response<Order>
+}
```
Bwvolleyball, line 38: Removing the `Response` wrapper to match

```suggestion
import org.junit.jupiter.api.Test
import retrofit2.Retrofit
import retrofit2.converter.gson.GsonConverterFactory
import retrofit2.converter.jackson.JacksonConverterFactory
import retrofit2.converter.moshi.MoshiConverterFactory
import retrofit2.http.Body
import retrofit2.http.GET
import retrofit2.http.POST
import retrofit2.http.Path

enum class OrderStatus { PENDING, COMPLETED, CANCELLED }

data class Order(val id: String, val status: OrderStatus)

interface OrderService {

    @GET("orders/{id}")
    suspend fun getOrder(@Path("id") orderId: String): Order

    @POST("orders")
    suspend fun createOrder(@Body order: Order): Order
}
```

PR #1040, opened 2024-08-02
[KTLN-843] Add samples

LeoColman:


Bwvolleyball:
retest this please

kotlin-testing/src/test/kotlin/com/baeldung/testresources/TestContainersResourceManagementUnitTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.extensions.testcontainers.JdbcDatabaseContainerExtension
+import io.kotest.matchers.shouldBe
+import org.testcontainers.containers.PostgreSQLContainer
+
+/**
+ * livetest. This class requires a docker environment to be setup on one's computer so that testcontainers can use
+ * it.
+ */
+class TestContainersResourceManagementLiveTest : StringSpec({
```
KevinGilmore, line 13: Shouldn't the file also be named the same as the class - ending with "LiveTest"? I know it's not strictly required in Kotlin but it would align with site guidelines.

PR #1041, opened 2024-08-02
[KTLN-449] Difference Between Lock.withLock and synchronized in Kotlin

albertache1998:
Added unit tests

### DIFF
--- a/core-kotlin-modules/core-kotlin-advanced-3/src/test/kotlin/com/baeldung/withLockVsSynchronize/WithLockVsSynchronizeUnitTest.kt
+++ b/core-kotlin-modules/core-kotlin-advanced-3/src/test/kotlin/com/baeldung/withLockVsSynchronize/WithLockVsSynchronizeUnitTest.kt
@@ -46,10 +46,6 @@ class WithLockVsSynchronizeUnitTest {
     fun synchronizedMethod() {
         println("Synchronized function call")
     }
-
-    mutexLock.withLock {
-        println("Synchronized function call")
-    }
 }
 
 class CounterClass {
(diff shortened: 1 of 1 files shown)


core-kotlin-modules/core-kotlin-advanced-3/src/test/kotlin/com/baeldung/withLockVsSynchronize/WithLockVsSynchronizeUnitTest.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.jupiter.api.Test
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.thread
+import kotlin.concurrent.withLock
+
+
+class WithLockVsSynchronizeUnitTest {
+
+    object ObjectToLockOn
+    val mutexLock = Mutex()
```
Bwvolleyball, line 16: I think this is unused, let's remove this.

```suggestion

```

core-kotlin-modules/core-kotlin-advanced-3/src/test/kotlin/com/baeldung/withLockVsSynchronize/WithLockVsSynchronizeUnitTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        println("Synchronized function call")
+    }
+
+    @Synchronized
+    fun synchronizedMethod() {
+        println("Synchronized function call")
+    }
+
+    fun performSynchronizedOperation(){
+        val lockObject = ReentrantLock()
```
Bwvolleyball, line 53: This lock needs to be declared outside of the function, otherwise it won't actually synchronize anything because we create a new lock with every function call.

PR #1077, opened 2024-08-22
[KTLN-571] Getting the Value of a Custom Property Defined in 'gradle.properties' in Kotlin

albertache1998:
Added unit tests


core-kotlin-modules/core-kotlin-advanced-3/pom.xml
```diff
@@ -71,6 +71,18 @@
...
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M7</version>
+                <configuration>
+                    <environmentVariables>
+                        <CUSTOM_PROPERTY>CustomValue</CUSTOM_PROPERTY>
+                        <API_URL>https://api.example.com</API_URL>
+                    </environmentVariables>
+                </configuration>
+            </plugin>
+
```
Bwvolleyball, line 85: I think this is leftover, and is probably not needed anymore?

kotlin-build-plugins/gradle.properties
```diff
@@ -0,0 +1,2 @@
+customProperty=CustomValue
+apiUrl=https://api.example.com
```
Bwvolleyball, line 2: generally, properties should be all lowercase and period separated.

```
custom.property
api.url
```

Bwvolleyball:
retest this please

core-kotlin-modules/core-kotlin-advanced-3/pom.xml
```diff
@@ -70,7 +70,6 @@
                     </execution>
                 </executions>
             </plugin>
-
```
KevinGilmore, line 73: No reason to change this file. Let's revert it.

PR #1127, opened 2024-10-16
KTLN-785 - Thread Safe Data Structures in Kotlin

hangga:


core-kotlin-modules/core-kotlin-concurrency-4/src/test/kotlin/com/baeldung/threadsafe/ThreadSafeUnitTest.kt
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    @Test
+    fun `test using Collections-synchronizedMap to prevent thread-safety issue`() {
+        val map = Collections.synchronizedMap(HashMap<Int, String>())
+
+        thread {
+            for (i in 1..100) {
+                map[i] = "Thread 1 - $i"
+                Thread.sleep(100) // simulate delay
```
theangrydev, line 253: Reduce the delay a bit so the test runs faster

core-kotlin-modules/core-kotlin-concurrency-4/src/test/kotlin/com/baeldung/threadsafe/ThreadSafeUnitTest.kt
```diff
@@ -0,0 +1,298 @@
...
+            }
+        }.join()
+
+        thread { map.remove(200) }.join()
+
+        assertEquals(199, map.size)
+    }
+
+    @Test
+    fun `test using ConcurrentHashMap for alternative to Collections-synchronizedMap`() {
```
theangrydev, line 269: This one doesn't make sense, if we are joining each time then it's basically single-threaded + it should be possible to modify concurrently with a ConcurrentHashMap without getting a ConcurrentModificationException

core-kotlin-modules/core-kotlin-concurrency-4/src/test/kotlin/com/baeldung/threadsafe/ThreadSafeUnitTest.kt
```diff
@@ -0,0 +1,298 @@
...
+
+        for (item in list) {
+            if (item == 3) {
+                list.remove(item)
+            }
+        }
+    }
+
+    @Test
+    fun `test using AtomicInteger to prevent race-condition`() {
```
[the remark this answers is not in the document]
hangga: Yeah, I think I'll make another scenario for this atomic.

hangga:
Hi @theangrydev I made some changes here. After this, I will update the article. (if you have approved this pr)

PR #1137, opened 2024-10-26
KTLN-575 Testing Flow in Kotlin

naoussi:
Added unit tests


core-kotlin-modules/core-kotlin-flows/pom.xml
```diff
@@ -32,6 +32,20 @@
...
+            <artifactId>kotlinx-coroutines-core</artifactId>
+            <version>1.6.0</version> <!-- Use the latest stable version -->
+        </dependency>
+
+        <!-- Kotlin Coroutines Test -->
+        <dependency>
+            <groupId>org.jetbrains.kotlinx</groupId>
+            <artifactId>kotlinx-coroutines-test</artifactId>
+            <version>1.6.0</version> <!-- Use the latest stable version -->
+            <scope>test</scope>
```
Bwvolleyball, line 46: Let's pull these versions out into the properties block below, and we probably don't need these comments to use the latest version.

core-kotlin-modules/core-kotlin-flows/src/test/kotlin/com/baeldung/flowTest/FlowTestUnitTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.assertThrows
+
+class FlowTestUnitTest {
+
+        @Test
+        fun `simpleFlow should emit 1 2 3`() = runTest {
+            val flow = simpleFlow().toList()
+
+            assertEquals(listOf(1, 2, 3), flow)
+        }
```
Bwvolleyball, line 21: Formatting might be a little off here.

PR #1155, opened 2024-11-30
Java 42077

Bipinkumar27:


pom.xml
```diff
@@ -489,6 +489,7 @@
 
                 <module>core-kotlin-companion</module>
                 <module>core-kotlin-modules</module>
```
anastasiosioannidis, line 491: Why did we add it there?

Bipinkumar27: Because only integration test cases were failing not default

PR #1166, opened 2024-12-09
KTLN-576 Create a List from Kotlin Flow 

naoussi:
Added unit test


core-kotlin-modules/core-kotlin-flows/src/test/kotlin/com/baeldung/flowToList/FlowToListUnitTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    fun `create list from flow using flatMapConcat method`() = runTest {
+        val flow = flowOf(1, 2, 3)
+        val result = flow.flatMapConcat { value ->
+            flowOf(value * 10)
+        }.toList()
+
+        assertEquals(listOf(10, 20, 30  ), result)
+    }
```
Bwvolleyball, line 54: We can safely remove this test from the PR.

PR #1181, opened 2025-01-16
feat(kotlin-collections): add list deep clone tests

husam-e:


core-kotlin-modules/core-kotlin-collections-2/src/test/kotlin/com/baeldung/lists/ListCopyUnitTest.kt
```diff
@@ -18,4 +18,44 @@ class ListCopyUnitTest {
...
+
+        val address1 = Address("Sesame Street", 1, "CartoonLand")
+        val address2 = Address("Sesame Street", 2, "CartoonLand")
+        val persons = listOf(Person("Bob", 20, address1), Person("Alice", 21, address2))
+
+        val personsCopy = persons.map { it.clone() }
+
+        assertThat(personsCopy).isNotSameAs(persons)
+        assertThat(personsCopy[0]).isNotSameAs(persons[0])
+        assertThat(personsCopy[0].address).isEqualTo(persons[0].address)
```
Bwvolleyball, line 55: Was this one intended to be the `isNotSameAs()` check?
```suggestion
        assertThat(personsCopy[0].address).isNotSameAs(persons[0].address)
```

husam-e: can add that as well, this is to show the data hasnt changed, but good call on also showing its a different reference

